PR #43, opened 2020-09-03
Add e2e tests for the playground app

lbalmaceda:
This PR is adding cypress integration tests for the playground application, testing:
- the login/logout flows
- silent auth token refresh
- protected routes redirect
- default playground options

I faced a caching issue with the cypress binary, documented [here](https://docs.cypress.io/guides/guides/continuous-integration.html#Caching). I couldn't make the orb cache multiple directories, and I'm not sure of the behavior of invoking the `save_cache` step twice. I refactored the script removing the orb usage so it looks more verbose now. I've raised an issue to discuss if the orb can be improved here `https://github.com/CircleCI-Public/node-orb/issues/53`.


.circleci/config.yml
```diff
@@ -1,25 +1,30 @@
...
+          command: ./node_modules/.bin/codecov
+      - store_artifacts:
+          path: projects/playground/e2e/videos
+      - store_artifacts:
+          path: projects/playground/e2e/screenshots
+      - store_test_results:
+          path: projects/playground/test-results
       - store_test_results:
           path: projects/auth0-angular/test-results
-workflows:
```
lbalmaceda, line 22: removing this section to rely on the "default workflow" which is named "build" instead. Will save us some lines here and will change the GH status check name. I'll update the settings after this PR is merged.

.circleci/config.yml
```diff
@@ -1,25 +1,30 @@
...
-                name: Upload coverage
-                command: ./node_modules/.bin/codecov
+      - restore_cache:
+          key: dependencies-{{ checksum "package-lock.json" }}
+      - run: npm ci
+      - save_cache:
+          key: dependencies-{{ checksum "package-lock.json" }}
+          paths:
+            - ~/.npm
+            - ~/.cache
```
lbalmaceda, line 15: had to move to this approach for the cypress binary to cache correctly

projects/playground/e2e/integration/playground.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+
+if (!EMAIL || !PASSWORD) {
+  throw new Error(
+    'You must provide CYPRESS_USER_EMAIL and CYPRESS_USER_PASSWORD environment variables'
+  );
+}
+
+const loginToAuth0 = () => {
+  cy.get('.auth0-lock-form')
+    .should('have.length.above', 1)
```
lbalmaceda, line 12: waits until the lock widget is displayed and has more than 1 children view, otherwise, it would try to find elements when the progress bar is still there.

projects/playground/e2e/integration/playground.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+    cy.get('#login').should('be.visible').click();
+    cy.url().should('include', 'https://brucke.auth0.com/login');
+    // logs in with the "last time you logged in with" button
+    cy.get('.auth0-lock-last-login-pane > a').should('be.visible').click();
+
+    cy.get('#logout').should('be.visible').click();
+    cy.get('#login').should('be.visible');
+  });
+
+  it('do federated logout (auth0 session is cleared)', () => {
```
stevehobbsdev: Do we want to keep this test given that we can't verify that we _actually_ do a federated logout?

If it's just to test that the options are specified, we have unit tests that cover that.

stevehobbsdev (commented):
LGTM, just left a comment about the federated logout.

projects/playground/e2e/integration/playground.spec.ts
lbalmaceda: I think there's value keeping it. Maybe we can rename the test to no longer make reference to the "federated" mode. We can even skip checking that checkbox and leaving both boxes unchecked. That would do a normal non-local logout. WDYT?

Checkboxes aside, what this test is asserting is that the auth0 session is cleared. How it does that is by checking that after logging out, visiting the login page will not remember who logged in previously.

This is how the "second login" looks for this test:
image

And this is how it looks for the "local logout" test:
image

stevehobbsdev: Yeah I think that's better suited to a normal non-local logout, I'd be happy with that change 👍

lbalmaceda: Done. I'm not "unchecking" any checkboxes because they are not checked by default. That default state is also tested in one of our first tests.

.circleci/config.yml
stevehobbsdev: I removed the "build-and-test" check requirement from the GitHub project 👍

stevehobbsdev (approved):

PR #44, opened 2020-09-05
docs: add return type to the example in the README

ahasall:
The example in the README creates a linting issue. The signature of the `loginWithRedirect` method should have a return type.


stevehobbsdev (approved):
Thanks @ahasall!

PR #440, opened 2023-03-25
Return jsdoc for `AuthClientConfig`

AlexElin:
Return jsdoc for `AuthClientConfig`.  
The type and its jsdoc were unintentionally separed in https://github.com/auth0/auth0-angular/pull/413


frederikprijck (approved):
Thanks!

PR #663, opened 2024-11-26
Improve the `provide` function in order to allow passing a factory function

LcsGa:
### Description

In some cases, we could need to use some other providers in order to configure auth0.
In this regard, I added the possibility to also provide a factory function and not just a basic value to allow this.

```ts
provideAuth0(() => {
    const someProvider = inject(SomeProvider);
    return {
        ...
        /* we can now use someProvider here */
    }
});
```

> This improvement doesn't include any breaking change.

### Testing

There is no existing unit tests for this provide function.

You can manually test it by playing with the playground, using this provide function and passing a factory function instead of an object.

### Checklist

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -1,14 +1,20 @@
 # Examples using auth0-angular
 
-- [Add login to your application](#add-login-to-your-application)
-- [Add logout to your application](#add-logout-to-your-application)
-- [Checking if a user is authenticated](#checking-if-a-user-is-authenticated)
-- [Display the user profile](#display-the-user-profile)
-- [Protect a route](#protect-a-route)
-- [Call an API](#call-an-api)
-- [Handling errors](#handling-errors)
-- [Organizations](#organizations)
-- [Standalone Components and a more functional approach](#standalone-components-and-a-more-functional-approach)
+- [Examples using auth0-angular](#examples-using-auth0-angular)
+  - [Add login to your application](#add-login-to-your-application)
+  - [Add logout to your application](#add-logout-to-your-application)
+  - [Checking if a user is authenticated](#checking-if-a-user-is-authenticated)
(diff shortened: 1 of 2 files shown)


LcsGa:
About the `Example.md` formatting changes, it seems that there is a git hook behind but I can't find it...
I guess that the new formatting isn't wanted, right?

tusharpandey13:
Hi @LcsGa, thank you for contributing to Auth0! 
Please note that we have acknowledged this request and will consider incorporating this into the codebase for the next minor release of auth0-angular. While we are aware of this request, since this is a new feature with no prior tests, we cannot provide an accurate timeline yet, but you can expect activity on this in January 2025.
Thanks.

Tiime-LcsGa:
Hi! any news on this?

yogeshchoudhary147:
Hi @LcsGa, sorry for not responding for so long, that was on our end. The idea here is valid and we would like to get this merged. To move forward we need a rebase on main as there are conflicts, unit tests for the new behaviour, the unrelated formatting changes in `EXAMPLES.md` to be removed, and the commits to be GPG signed. If you are still interested please let us know and we will be happy to review.

yogeshchoudhary147:
Hi @LcsGa, apologies for the delay and thanks for the contribution!

After reviewing, we are closing this as the use case is already covered by the documented [Dynamic configuration](https://github.com/auth0/auth0-angular#dynamic-configuration) pattern using `provideAppInitializer` + `AuthClientConfig.set()`:

```ts
provideAuth0(),
provideAppInitializer(() => {
  const someProvider = inject(SomeProvider);
  inject(AuthClientConfig).set({
    domain: someProvider.domain,
    clientId: someProvider.clientId,
    // ...
  });
}),
```

Since this already handles the need, we would prefer to keep `provideAuth0`'s signature simple.

PR #767, opened 2026-01-12
feat: add connectAccountWithRedirect method to AuthService

yogeshchoudhary147:
## Description

Adds support for the `connectAccountWithRedirect()` method to auth0-angular, exposing the Connect Account functionality already available in auth0-spa-js.

## Changes

- ✨ Added `connectAccountWithRedirect()` method to `AuthService` that wraps the auth0-spa-js client method
- 📤 Exported `RedirectConnectAccountOptions` type from `public-api.ts` 
- 📝 Added Connect Accounts usage example to `EXAMPLES.md` with configuration and implementation details
- ✅ Added unit tests for the new method with mock setup and test cases

## Motivation

The `connectAccountWithRedirect()` method is available in auth0-spa-js but was not exposed in auth0-angular. This prevented Angular customers from using the Connect Account flow with MRRT (Multi-Refresh Token Rotation).

Without this method, customers were forced to use silent login (iframe-based) as a workaround, which fails in Safari and other browsers with strict third-party cookie policies.

## Implementation Details

- Follows the same Observable-based pattern as `loginWithRedirect()`
- Returns `Observable<void>` consistent with other redirect methods
- The existing `handleRedirectCallback()` method already supports `ConnectAccountRedirectResult`, so no additional changes needed for callback handling
- Method includes comprehensive JSDoc documentation with examples

## Testing

- Added unit tests that verify the method calls the underlying auth0-spa-js client
- Tests cover both basic usage and all optional parameters (scopes, authorization_params, redirectUri, appState, openUrl)

PR #817, opened 2026-02-16
feat: add connected accounts callback handling

yogeshchoudhary147:
https://github.com/user-attachments/assets/a25c6039-77bb-4845-a09e-de6da046b45d

## Description
Completes the Connected Accounts implementation started in #767. Previously the SDK could initiate the connect flow but couldn't process the response when users returned from connecting their account. This PR adds the missing callback logic.

## Changes
- Add callback handling for the connected accounts redirect flow
- Expose connected account details via `appState$` after redirect completes
- Export related types: `ConnectedAccount`, `ResponseType`, `ConnectError`, `ConnectAccountRedirectResult`
- Update EXAMPLES.md with callback handling and listing connected accounts examples

## Note
The SPA SDK's role in Token Vault covers the Connected Accounts flow: initiating and completing account connections and managing them via the My Account API. Actual token retrieval from Token Vault requires a confidential backend client.

## Testing
4 new tests added, 131 total passing. E2E flow verified (see demo above).


projects/auth0-angular/src/lib/auth.service.ts
```diff
@@ -390,9 +391,17 @@ export class AuthService<TAppState extends AppState = AppState>
...
 
+        // Add response_type to appState
+        appState.response_type = response_type;
+
+        // If this is a connect account flow, add the connected account data to appState
+        if (response_type === ResponseType.ConnectCode) {
+          appState.connectedAccount = rest as ConnectedAccount;
+        }
+
         if (appState) {
```
gyaneshgouraw-okta: @yogeshchoudhary147 On line number 398, we have added 
`appState.response_type = response_type;`

This will always make the `appState` always truthy, even for non-connected account flow.
Will that not cause side effects ?

yogeshchoudhary147: Good catch! Removed the {} default so we no longer emit to `appState$` when no `appState` was provided in a regular login flow.

gyaneshgouraw-okta (approved):

PR #893, opened 2026-04-09
fix(security): run npm ci with --ignore-scripts

jcchavezs:
This would avoid malicious scripts in vulnerable packages to be executed on CI

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -100,7 +100,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Install dependencies
-        run: npm ci
+        run: npm ci --ignore-scripts
 
       - name: Publish release to NPM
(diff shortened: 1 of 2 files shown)


yogeshchoudhary147:
Thanks for the contribution! This makes sense for hardening our CI pipeline against supply chain attacks.

Could you also add `--ignore-scripts` to the other two `npm ci` calls that were missed?

1. `.github/actions/build/action.yml` (line 21)
2. `.github/actions/npm-publish/action.yml` (line 33 — `npm ci --include=dev` → `npm ci --include=dev --ignore-scripts`)

That way we're consistent across the entire CI surface.

yogeshchoudhary147:
Closing in favour of #897 which adds `ignore-scripts=true` via `.npmrc`, covering all `npm install` invocations rather than just CI steps.

PR #934, opened 2026-07-06
feat: add support for Online Access (Online Refresh Tokens)

NandanPrabhu:
## Summary

Adds full support for **Online Access (Online Refresh Tokens / ORTs)** in `auth0-angular`, mirroring the underlying `auth0-spa-js` feature.

Most of the configuration surface (`refreshTokenMode`, `useDpop`, `useMrrt`, `useRefreshTokens`) already passed through automatically since `AuthConfig extends Auth0ClientOptions` and `Auth0ClientFactory` spreads config directly — no changes needed there. 

## Changes

- Exported `RefreshTokenMode`, `InvalidConfigurationError`, `MissingScopesError` from `public-api.ts`
- New "Online Access (Online Refresh Tokens)" section in `EXAMPLES.md`, covering both `AuthModule.forRoot` and standalone `provideAuth0` configuration styles
- Tests added to auth.client.spec.ts for auth client if refresh token mode is online 
- Tests added to auth.service.spec.ts for testing logout behaviour incase of ORT 

## Usage

```ts
import { AuthModule, RefreshTokenMode } from '@auth0/auth0-angular';

AuthModule.forRoot({
  domain: 'YOUR_AUTH0_DOMAIN',
  clientId: 'YOUR_AUTH0_CLIENT_ID',
  useRefreshTokens: true,
  refreshTokenMode: RefreshTokenMode.Online,
  useDpop: true,
  authorizationParams: { redirect_uri: window.location.origin },
});
```

## Testing

- [x] All tests pass (`npx jest --no-coverage` — 184 tests)
- [x] `npx tsc --noEmit` — zero TypeScript errors
- [x] `npx ng build auth0-angular` — builds cleanly
- [x] `npx ng lint` — zero errors
- [x] Built against a local, linked copy of `auth0-spa-js` and smoke-tested `RefreshTokenMode` resolve correctly


NandanPrabhu:
Snyk vulnerability check failing — root cause is an Angular major version, not a patch fix

  CI is flagging 4 vulnerabilities in @angular/common, @angular/compiler, and @angular/core (currently pinned to ^19.2.21, resolving to 19.2.25):

  - @angular/common@19.2.25: ReDoS (High) — SNYK-JS-ANGULARCOMMON-17356486 (https://security.snyk.io/vuln/SNYK-JS-ANGULARCOMMON-17356486)
  - @angular/common@19.2.25: Use of Weak Hash (Critical) — SNYK-JS-ANGULARCOMMON-17356555 (https://security.snyk.io/vuln/SNYK-JS-ANGULARCOMMON-17356555)
  - @angular/compiler@19.2.25: XSS (High) — SNYK-JS-ANGULARCOMPILER-17356585 (https://security.snyk.io/vuln/SNYK-JS-ANGULARCOMPILER-17356585)
  - @angular/core@19.2.25: Modification of Assumed-Immutable Data (High) — SNYK-JS-ANGULARCORE-17353317 (https://security.snyk.io/vuln/SNYK-JS-ANGULARCORE-17353317)

  19.2.25 is already the latest published patch in the 19.x line — there is no 19.x release that fixes these. Snyk's own remediation advice confirms the fix only lands in
  20.3.25, i.e. this requires bumping across the Angular 19 → 20 major version boundary.

  I tried a narrow fix — bumping only @angular/common, @angular/compiler, and @angular/core to ^20.3.25 while leaving the rest of the @angular/* packages at 19.2.21 — but npm install fails with ERESOLVE: @angular/animations, @angular/platform-browser, @angular/router, @angular/forms, and @angular/compiler-cli all declare a peer dependency  on @angular/common@"19.2.25" (and its siblings) at that exact version. Angular's own packages are lockstep-versioned, so they can't be upgraded independently — this isn't fixable without upgrading the full Angular toolchain together.

  This means fixing the Snyk finding requires a full Angular 19 → 20 major upgrade (@angular/*, @angular/cli, @angular-devkit/build-angular, @angular-eslint/*, ng-packagr,
  run via ng update), which carries real breaking-change risk for this repo specifically:
  - This repo is still on RxJS 6 (^6.6.7) — Angular 20 likely requires RxJS 7+, which is itself a breaking change for any internal Observable usage.
  - Angular 20 raises the minimum supported Node.js version — CI/build images may need updating.
  - ng update runs automated codemods but the playground app, schematics, and any Angular APIs this SDK depends on need to be re-verified afterward (full rebuild + unit +
  e2e suite).

EXAMPLES.md
```diff
@@ -820,6 +820,117 @@ bootstrapApplication(AppComponent, {
 });
 ```
 
+## Online Access (Online Refresh Tokens)
```
cschetan77, line 824: Update the TOC.

EXAMPLES.md
```diff
@@ -820,6 +820,117 @@ bootstrapApplication(AppComponent, {
...
+
+try {
+  // AuthModule.forRoot / provideAuth0 constructs the client during app bootstrap.
+  // Wrap it, or validate your configuration up front, to catch misconfiguration.
+} catch (e) {
+  if (e instanceof InvalidConfigurationError) {
+    console.error(e.error_description); // includes the suggested fix
+  }
+}
+```
```
cschetan77: The `try/catch` in the example won't catch the error in practice. `Auth0Client` is constructed lazily by Angular's DI system, the factory runs when the first component that consumes AuthService is bootstrapped, which happens inside the Promise returned by `bootstrapApplication()` (or inside NgModule bootstrap). By the time you write `provideAuth0({...})` or `AuthModule.forRoot({...})`, no client has been constructed yet, the factory hasn't run. So a synchronous try/catch block around the call site is dead code. 

The `InvalidConfigurationError` surfaces at component bootstrap time, inside the Promise Angular returns. There are two valid ways to handle it:
Option 1: `.catch()` on the bootstrap Promise (simplest, catches it at startup):
```js
  bootstrapApplication(AppComponent, {
    providers: [
      provideAuth0({
        domain: 'YOUR_DOMAIN',
        clientId: 'YOUR_CLIENT_ID',
        useRefreshTokens: true,
        refreshTokenMode: RefreshTokenMode.Online,
        useDpop: true,
        authorizationParams: { redirect_uri: window.location.origin },
      }),
    ],
  }).catch((e) => {
    if (e instanceof InvalidConfigurationError) {
      console.error('Auth0 misconfiguration:', e.error_description);
    }
  });
```

  Option 2: Angular `ErrorHandler` (catches it in the running app, same pattern as React's error boundary):
```js
  import { ErrorHandler, Injectable } from '@angular/core';
  import { InvalidConfigurationError } from '@auth0/auth0-angular';

  @Injectable()
  export class AppErrorHandler implements ErrorHandler {
    handleError(error: unknown): void {
      if (error instanceof InvalidConfigurationError) {
        console.error('Auth0 misconfiguration:', error.error_description);
        return;
      }
      console.error(error);
    }
  }

  // in app.config.ts:
  { provide: ErrorHandler, useClass: AppErrorHandler }
```
Angular calls `ErrorHandler.handleError` with the original InvalidConfigurationError (not wrapped) and also rejects the bootstrap Promise with it.

EXAMPLES.md
```diff
@@ -820,6 +820,117 @@ bootstrapApplication(AppComponent, {
...
+
+- **Session-bound** — valid only while the underlying Auth0 session is active. When the session ends (logout, idle/absolute session expiry, or an admin revoking the session), the ORT stops working.
+- **Non-rotating** — refreshing an access token with an ORT does **not** issue a new refresh token; the same ORT is reused for the life of the session.
+
+Read more about [Online Refresh Tokens](https://auth0.com/docs/secure/tokens/refresh-tokens/online-refresh-tokens/online-refresh-tokens) to decide whether this fits your application.
+
+> [!IMPORTANT]
+> Online access requires DPoP. Sender-constraining the token via [DPoP](#device-bound-tokens-with-dpop) is mandatory because the ORT is non-rotating — binding it to the browser's key pair is what mitigates token replay if it is exfiltrated. You must set `useDpop: true` explicitly; the SDK does not enable it for you.
+>
+> This also requires the `online_refresh_tokens` flag to be enabled for your Auth0 tenant, and `allow_online_access` to be enabled on the resource server you log in with (on by default).
```
cschetan77: We sure `allow_online_access` is on by default ? 
Also, I couldn't find `online_refresh_tokens` flag, we sure this exist and have to be enabled!

NandanPrabhu: this needs to be removed

EXAMPLES.md
NandanPrabhu: Good catch, added the missing TOC entry right after the DPoP one:

`- [Online Access (Online Refresh Tokens)](#online-access-online-refresh-tokens)`

EXAMPLES.md
NandanPrabhu: You're right, this was dead code. `Auth0Client` is constructed lazily by Angular's DI system, so `InvalidConfigurationError` surfaces during app bootstrap, not at the `AuthModule.forRoot`/`provideAuth0` call site. Replaced the synchronous `try/catch` with `.catch()` on the `bootstrapApplication` promise, and added an `ErrorHandler`-based alternative for catching it in the running app too.

EXAMPLES.md
NandanPrabhu: Fair point on both — I couldn't verify `online_refresh_tokens` as an actual tenant flag either, so removed that reference entirely (the Early Access note above already covers gating). Also dropped the unverified "(on by default)" claim for `allow_online_access` since I couldn't confirm that from source. Left `allow_online_access` itself since that's a real resource-server setting referenced in auth0-spa-js.

cschetan77 (approved):

PR #956, opened 2026-07-17
chore: pin eslint-plugin-import, jsdoc, and prefer-arrow versions

yogeshchoudhary147:
## Summary

- Replaces `"latest"` version specifiers for `eslint-plugin-import`, `eslint-plugin-jsdoc`, and `eslint-plugin-prefer-arrow` with pinned semver ranges (`^`)
- Updates `package-lock.json` to sync `eslint-plugin-jsdoc` from `63.0.13` → `63.0.14`

Using `"latest"` causes `npm ci` to fail whenever a new release drops before the lock file is updated, since `npm ci` requires `package.json` and `package-lock.json` to be in sync.

## Test plan

- [x] CI passes (`npm ci` succeeds on all jobs)
- [x] Lint, build, and unit tests pass


developerkunal (approved):
LGTM

